Issue #8, opened 2015-12-04
Bug in implementation of `decode` for Either


axman6:
At https://github.com/well-typed/binary-serialise-cbor/blob/master/Data/Binary/Serialise/CBOR/Class.hs#L202 the recent PR to add more strictness (#5) failed to add in the `Left` and `Right` constructors when decoding `Either`s. I haven't tested it, but I expect that at best this will always cause decoding to fail, and at worse could cause an infinite loop, but that's probably unlikely (the instance for `Either` is relying on its own instance recursively, the x's both have type `Either a b`)

adamgundry:
Well spotted, thanks!

PR #9, opened 2015-12-14
Further strictness annotations

adamgundry:
This follows up on my previous PR (#5), applying the same treatment to `decodeTerm`. Also includes a fix for #8.

### DIFF
--- a/Data/Binary/Serialise/CBOR/Term.hs
+++ b/Data/Binary/Serialise/CBOR/Term.hs
@@ -29,6 +29,7 @@ import qualified Data.ByteString as BS
 import qualified Data.ByteString.Lazy as LBS
 import           Data.Monoid
 import           Control.Applicative
+import           Control.Monad
 
 import Prelude hiding (encodeFloat, decodeFloat)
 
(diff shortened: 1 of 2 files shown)


Data/Binary/Serialise/CBOR/Term.hs
```diff
@@ -102,7 +102,8 @@ decodeTerm :: Decoder Term
 decodeTerm = do
     tkty <- peekTokenType
     case tkty of
-      TypeUInt   -> fromWord <$> decodeWord
+      TypeUInt   -> do w <- decodeWord
+                       return $! fromWord w
```
axman6: If it does't already exist, why not make a combinator for this:

```haskell
(<$!>) :: Monad f => (a -> b) -> f a -> f b
g <$!> fa = do a <- fa; return $! g a
```
Then the changes would be minor.

Also, would there be any advantage in making it `!a <- fa...`?

axman6: Seems Data.Monad exports exactly this function since 4.8.0.0 (GHC 7.10 iirc).

adamgundry: True, I suppose we could conditionally define `<$!>` locally (we need to support GHC back to 7.8 at least). I don't think there's any need to force the `a` if `g` is strict, which it usually is, but I guess it makes a small difference in the lazy case.

PR #36, opened 2016-03-10
Remove intermediate List in decoding vectors

arianvp:
We converted to a list before converting to vector when decoding (unboxed) vectors. This removed the intermediate step.


arianvp:
Closed in favour of https://github.com/well-typed/binary-serialise-cbor/pull/38

PR #45, opened 2016-04-02
Removed MINIMAL pragma from class definition.

### LINKED ISSUE #44: Remove the MINIMAL pragma from Serialise
Given that there are `Generic` default methods for the `Serialise` typeclass, the MINIMAL pragma is useless and only serves to generate warnings (if there isn't a `Generic` instance, it will fail with type error). Same problem was with aeson: https://github.com/bos/aeson/issues/290 or https://ghc.haskell.org/trac/ghc/ticket/10959

ondrap:
Removes the MINIMAL pragma from `Serialise` class - when `Generic` default methods are defined, it doesn't work as intended and only generates warning.
Closes #44


thoughtpolice:
Good catch! Will integrate soon.

PR #57, opened 2016-05-09
Fix warnings

michalt:
There's been a couple of unnecessary imports in the main package and a
bunch of various warnings in the benchmarking code. This fixes almost
all of them (except for bench/versus/Macro/CBOR where I decided to simply
disable the warning about missing signatures - it doesn't seem to be
that valuable in this case and would require a fair amounut of
boilerplate).

### DIFF
--- a/Data/Binary/Serialise/CBOR/Read.hs
+++ b/Data/Binary/Serialise/CBOR/Read.hs
@@ -31,6 +31,10 @@ import qualified Data.Binary.Get as Bin (Decoder(..))
 #if !MIN_VERSION_base(4,8,0)
 import           Control.Applicative
 #endif
+#if defined(ARCH_32bit)
+import           GHC.Int
+#endif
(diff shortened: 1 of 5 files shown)


thoughtpolice:
The AppVeyor failure is transient but I'd like to wait and restart it to make sure we don't break the 32 bit paths, after I so lovingly got it working. I'll take a look a bit later today

thoughtpolice:
Aha! As I suspected, this broke the 32 bit build: https://ci.appveyor.com/project/thoughtpolice/binary-serialise-cbor/build/1.0.158 :)

michalt:
Oops. It's also missing some of the Prelude re-exports from base-4.8. I'll ping you once everything looks ok in CI.

michalt:
@thoughtpolice seems like everything works now and is ready for review :-)

PR #85, opened 2016-06-23
Provide a maximally simple example of serialising

mikedlr:
This is a maximally simple example of Data.Binary.Serialise.CBOR  usage to make it easy for people to start from.

### DIFF
--- a/demos/Simple.hs
+++ b/demos/Simple.hs
@@ -1,0 +1,15 @@
+-- This is indended to be about the simplest possible use of
+-- Data.Binary.Serialise.
+
+import Data.Binary.Serialise.CBOR as CBOR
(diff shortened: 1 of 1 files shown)


thoughtpolice:
Sorry about missing this! Thanks - this looks fine, although I'll probably do some minor touchups on the style, etc. And it would be good to have a similar example somewhere in the tutorial when I write it later...

axman6:
I feel it would at least be nice if the demo at least read the data back in. I'm not sure this really adds anything that 3 minutes looking at the API couldn't have told you already - it should at the very least include an example of writing an instance for Serialise for a user defined type.

bgamari:
Perhaps this should be merged?

PR #87, opened 2016-07-12
demos: fix unhandled token in Aeson demo

jb55:
I was messing around with the Aeson demo and I noticed it was hitting `unexpected CBOR token type for a JSON value: TypeListLenIndef` every time. Not sure if this is the optimal fix but it works.


thoughtpolice:
Thanks! I'll take a closer look at this soon; sorry for the delay.

PR #117, opened 2017-06-21
Restructure modules

### LINKED ISSUE #104: Consolidate types into a single module.
In the quest to remove all boot files, because boot files are Evil, Bad, and Wrong, I think the best thing to do would simply be to move all the definitions of the major types into a single `.Types` module, much how `Class.hs` is the one place for class instances, too.

I'll be taking a whack at this later, but this is just for notice.

bgamari:
Here we resolve the import loop between Encoding and FlatTerm by moving their
types to dedicated modules, allowing FlatTerm to import Encoding.Types, enabling
it to define toFlatTerm. Encoding can then import FlatTerm.toFlatTerm for use
in the (now orphaned) Show Encoding instance. Despite the orphan, I think this
is the best we can do since FlatTerm and Encoding can't be defined in the same
source file due to overlapping constructor names.

Fixes #104. Depends upon #116.


axman6:
The comment about not having these types in the same module struck me as odd, is there any reason these two types need to have constructors with the same name?

bgamari:
I'm not sure I see why we would *want* them to be in the same module. `FlatTerm` and `Encoding` are quite orthogonal and I consider it to be a feature that the module structure reflects that.

However, even if we *did* want to place the types in the same module I would be very reluctant to do this sort of renaming at this stage given that it would bring few benefits and many costs (for both the maintainers of `cborg` as well as its users).

bgamari:
I tend to agree. I think the code is generally more readable prior to this change.

PR #119, opened 2017-06-22
travis: Bracket new-test with travis_wait to prevent timeouts

bgamari:


bgamari:
Well, things still timeout but this is at least not due to lack of output.

PR #133, opened 2017-07-31
binary-serialise-cbor: Rip out compatibility with GHC 7.8

bgamari:
Cabal check won't allow my conditional export trick and I don't think the
compatibility is worth the cost for a wrapper library.


bgamari:
This was merged.

PR #147, opened 2017-09-01
Add decoding variants for integers and floats that check canonicity of input

arybczak:
That's not 100% done yet (more info below), but I'm just putting it out for preliminary review and discussion.

This PR adds decoding variants that check whether representation of appropriate input value is canonical and fail if it's not.

Summary:
1) Canonicity checks for bounded integers are as described in https://tools.ietf.org/html/rfc7049#section-3.9.

2) For floating point numbers RFC suggests to choose the smallest representation that doesn't result in a loss of precision. However, that would require doing a lot of conversions on both encoding and decoding side, so I opted for simpler solution, namely:
* Canonical representation of a 16-bit float is always 2 bytes long (+ tag byte). If it's a NaN, it's always represented as `0xf97e00` (as per RFC).
* Canonical representation of a 32-bit float is always 4 bytes long (+ tag byte), except when it's a NaN, then it's `0xf97e00`.
* Canonical representation of a 64-bit float is always 8 bytes long (+ tag byte), except when it's a NaN, then it's `0xf97e00`.

3) For Integer, we consider canonical representation the following:
* If the number fits into 8 bytes, then it's represented as `Int` and `Int`-specific rules of canonicity apply.
* If it doesn't, then we require that unbounded binary representation is the shortest possible (i.e. its leading byte is not 0x00). Because of the first rule we also require that this representation is at least 9 bytes long.

There are currently two things that require some fiddling:
1) A few tests fail because handling of NaNs changed.
2) Canonicity of `Integer` is always checked even when we use non-canonical decoding variant (because we need information that is available only during parsing its representation). It's calculated lazily though, so I'm not sure how big of an issue that is.


axman6:
Could most of this be implemented by just using newtypes: 
```haskell
newtype CanonicalFloat16 = CF16 Float16

instance Serialise CanonicalFloat16 where...
```

Assuming we can use the safe coercions mechanisms, then you don't even need to change all your types to ensure that the canonical parsers are used, but you get the benefits. Just my 2¢

arybczak:
That is doable for all types but floats and integers (as they need access to the internal ByteString), but its usage would be really verbose (less so with TypeApplications, but still).

arybczak:
Alright, I fixed tests that were failing due to different handling of NaNs (note that #150 needs to be merged first for all tests to pass).

arybczak:
@dcoutts ping

PR #151, opened 2017-09-15
fix haddock link: Codec.CBOR.Tutorial -> Codec.Serialise.Tutorial

elaforge:
Like the subject says.


bgamari:
Great catches! Thanks for the fixes.

PR #154, opened 2017-11-06
Export GSerialiseProd(..) and GSerialiseSum(..).

ts468:


thoughtpolice:
Thanks! Would you mind removing the `.nix` files? I don't believe they're intentional, and they're rather unrelated to this change.

In general I'd rather have people use `cabal2nix` to vendor their own `.nix` files downstream, if they need to depend on this library (like we do at `$WORK`). That said, something in place for `nix-shell` would be nice...

ts468:
@thoughtpolice, thanks! Yes, the `.nix`-files were unintentionally added to this PR, and should be removed now.

thoughtpolice:
Thanks!

PR #155, opened 2017-11-07
Read: Maintain no-empty-chunks invariant of lazy ByteString

bgamari:
`Data.ByteString.Lazy.ByteString` expects the invariant that no chunk is empty.
However, runIDecode was breaking this when producing the remaining
content. Fix this.


thoughtpolice:
LGTM.

PR #159, opened 2018-01-03
Make DeserialiseFailure derive Eq

NunoAlexandre:
This makes it cleaner to compare instances of this type or of other
data structures holding a field of this type.


bgamari:
Sounds good to me.

NunoAlexandre:
Thanks.

PR #179, opened 2018-09-17
Make it build/work with ghc 8.4.3

erikd:


erikd:
Wow, making this compile with all the old versions was a bit of a pain in the neck. Managed to get ghc 7.10.3 working, but 7.8.4 seems to have been broken for some time.

thoughtpolice:
I think #180 will superscede this, but I haven't had a chance to test it all out. I'll get to it shortly.

quasicomputational:
If this PR's good to go, merging won't cause any problems with #180. If anything, that would make reviewing #180 easier because it'll reduce the patch size.

erikd:
@quasicomputational If this is merged first, then your PR would have to be rebased before it can be merged.

bgamari:
Let's just merge this.

PR #180, opened 2018-09-23
GHC 8.6 compatibility

quasicomputational:
This is based on #179.

It's not actually going to pass CI just yet because `th-expand-syns` needs a new release (but DanielSchuessler/th-expand-syns#11 has been merged, so hopefully that'll happen any moment now).

Once this is merged, a release would be very handy, please.


DanielSchuessler:
@quasicomputational: I did as @hvr suggested and added a metadata revision on Hackage; hope that works for you (might have to retrigger a Travis build).

serialise/bench/versus/Macro/DeepSeq.hs
```diff
@@ -1,9 +1,13 @@
+{-# LANGUAGE CPP #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 module Macro.DeepSeq where
 
 import Macro.Types
 
 import Control.DeepSeq
+#if MIN_VERSION_base(4,9,0)
```
hvr: shouldn't this rather be `MIN_VERSION_deepseq`? the `rnf1`/`rnf2` availability is not a proper function of the `base` version

hvr:
@bgamari @thoughtpolice while there's already build-plans for GHC 8.6 (see https://matrix.hackage.haskell.org/package/serialise) the relaxations for the library components requested here can be safely performed as metadata revisions (it's mostly about accepting `containers-0.6`)

if you don't have time, do you want me to perform those revisions to expedite things and unblock @quasicomputational ?

quasicomputational:
FWIW, revisions wouldn't really help me for a few reasons (I'm already using HEAD, and the GHC 8.6 upgrade plan is blocked elsewhere for me). It might benefit other users, though.

bgamari:
Thanks for doing this!

PR #188, opened 2018-11-09
serialise: Codec.Serialise.Class exports encodeContainerSkel, encodeM…

fumieval:
…apSkel and decodeMapSkel

I find them quite useful when defining an instance for my own Map-like datatype.


thoughtpolice:
Sorry this was missed for so long. I think this is alright to go in, though maybe @bgamari or @dcoutts have opinions about this as well.

bgamari:
Indeed, sorry for the long latency @fumieval. I would be okay with exposing these but we should really add some documentation to them (and perhaps place them under their own heading in the Haddocks). Could you take care of this?

fumieval:
Added some haddock comments.

PR #203, opened 2019-06-26
primitive-0.7.0.0 support

mitchellwrosen:
This patch bumps the `primitive` upper bound from `< 0.7` to `< 0.7.1`

### DIFF
--- a/serialise/serialise.cabal
+++ b/serialise/serialise.cabal
@@ -71,7 +71,7 @@ library
     ghc-prim                >= 0.3.1.0 && < 0.6,
     half                    >= 0.2.2.3 && < 0.4,
     hashable                >= 1.2     && < 2.0,
-    primitive               >= 0.5     && < 0.7,
+    primitive               >= 0.5     && < 0.8,
     text                    >= 1.1     && < 1.3,
     unordered-containers    >= 0.2     && < 0.3,
     vector                  >= 0.10    && < 0.13
--- a/cborg/cborg.cabal
+++ b/cborg/cborg.cabal
@@ -82,7 +82,7 @@ library
     deepseq                 >= 1.0     && < 1.5,
     ghc-prim                >= 0.3.1.0 && < 0.6,
     half                    >= 0.2.2.3 && < 0.4,
-    primitive               >= 0.5     && < 0.7,
+    primitive               >= 0.5     && < 0.8,
     text                    >= 1.1     && < 1.3
 
   if flag(optimize-gmp)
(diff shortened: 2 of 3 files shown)


serialise/serialise.cabal
```diff
@@ -69,7 +69,7 @@ library
     ghc-prim                >= 0.3.1.0 && < 0.6,
     half                    >= 0.2.2.3 && < 0.4,
     hashable                >= 1.2     && < 2.0,
-    primitive               >= 0.5     && < 0.7,
+    primitive               >= 0.5     && < 0.7.1,
```
phadej: why `<0.7.1` and not `<0.8` ?

phadej:
Screenshot from 2019-06-27 23-48-19

mitchellwrosen:
@phadej It definitely could be `< 0.8`, in fact it would probably cause less maintenance burden. I just put 0.7.1 to be conservative.

And about your screenshot, I'm not totally sure what I'm looking at here?

phadej:
@mitchellwrosen it's an evidence that `store` maintainers do lousy job :(

phadej:
about `<0.7.1` vs `<0.8` please read through https://pvp.haskell.org/

Screenshot from 2019-06-27 23-56-31

mitchellwrosen:
@phadej I'm aware of the PVP (the gist of it anyway) - again, I was just being conservative. There might be a reason why a maintainer would prefer to only allow patch bumps without noticing build failures (in case they don't trust upstream to properly follow the PVP themselves or something).

Anyway, overall I agree with you and am happy to bump to `< 0.8` if @thoughtpolice would prefer that.

And about `store` - sorry, I still don't follow - what about it? I see that neither `store` nor `cborg` depend on each other so what's the relevance here? (Just asking again in case I'm missing something pertinent to this patch).

phadej:
`serialise`s (which is in this repository) benchmarks depend on `store`.

mitchellwrosen:
Oh, right, you're addressing the CI failures. Ok, I'm all caught up, thanks :)

phadej:
I cannot restart the jobs, but now `store` should build fine.

boj:
I came here to essentially do the same thing. The bump to `< 0.8` does seem to be necessary, I can't get certain libaries to play nice in GHC 8.6.5 with this patch.

chessai:
I also would like this to go through. Thanks

mitchellwrosen:
Sorry @boj, I did not see your comment until just now. I've updated the bounds to `< 0.8`.

dcoutts:
To fix CI probably need to rebase once #206 is merged.

mitchellwrosen:
@dcoutts Rebased

PR #212, opened 2019-09-12
Fix a `SomeTypeRep` tagging issue

deepfire:
The demonstrator test fails:
```
type: SomeTypeRep
        cbor roundtrip:                       FAIL
          *** Failed! Exception: 'DeserialiseFailure 2 "Codec.CBOR.Class.decodeSomeTypeRep: unexpected tag\n"' (after 1 test):
          [Either (Maybe Int) (Proxy (Kind *) ('Type * [Char]))]
          Use --quickcheck-replay=754269 to reproduce.
        flat roundtrip:                       FAIL
          *** Failed! Falsified (after 1 test):
          [Either (Maybe Int) (Proxy (Kind *) ('Type * [Char]))]
          Use --quickcheck-replay=127810 to reproduce.
        flat term is valid:                   OK
          +++ OK, passed 100 tests.
```

### DIFF
--- a/serialise/src/Codec/Serialise/Class.hs
+++ b/serialise/src/Codec/Serialise/Class.hs
@@ -1216,11 +1216,11 @@ decodeSomeTypeRep = do
     case tag of
       0 | len == 1 ->
               return $! SomeTypeRep (typeRep :: TypeRep Type)
-      1 | len == 2 -> do
+      1 | len == 3 -> do
               !con <- decode
               !ks <- decode
               return $! SomeTypeRep $ mkTrCon con ks
-      2 | len == 2 -> do
+      2 | len == 3 -> do
               SomeTypeRep f <- decodeSomeTypeRep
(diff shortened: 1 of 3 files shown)


deepfire:
cc @thoughtpolice @dcoutts

serialise/src/Codec/Serialise/Class.hs
```diff
@@ -1248,7 +1248,9 @@ decodeSomeTypeRep = do
                       Just HRefl -> return $! SomeTypeRep $ Fun arg res
                       Nothing -> failure "Kind mismatch" []
                 Nothing -> failure "Kind mismatch" []
-      _ -> failure "unexpected tag" []
+      _ -> failure "unexpected tag"
+           [ "Tag: " ++ show tag
+           , "Len: " ++ show len ]
```
thoughtpolice: This is excellent, thanks!

PR #239, opened 2020-06-03
Implement TermF for Codec.CBOR.Term, for use with recursion schemes

infinity0:
I'm writing a mini UI library for browsing various data types, based on recursion schemes. But I think it's more natural for this definition to go in here, rather than my library defining similar things for lots of different data libraries (and then having to depend on all of them).

### DIFF
--- a/cborg/src/Codec/CBOR/Term.hs
+++ b/cborg/src/Codec/CBOR/Term.hs
@@ -1,5 +1,6 @@
-{-# LANGUAGE CPP          #-}
-{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP               #-}
+{-# LANGUAGE BangPatterns      #-}
+{-# LANGUAGE DeriveTraversable #-}
 
 -- |
 -- Module      : Codec.CBOR.Term
@@ -31,6 +32,8 @@ module Codec.CBOR.Term
   ( Term(..)    -- :: *
   , encodeTerm  -- :: Term -> Encoding
   , decodeTerm  -- :: Decoder Term
+  , TermF(..)   -- :: *
+  , termToF     -- :: (TermF a -> a) -> Term -> a
   ) where
 
 #include "cbor.h"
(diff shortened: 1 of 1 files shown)


infinity0:
Actually, I just found out that `recursion-schemes` has a TH utility that generates nearly the same code. So I could replace it with that, but then this library would have that additional extra dependency. (Inlining is not possible because the generated code defines instances of typeclasses defined in that library.)

cborg/src/Codec/CBOR/Term.hs
```diff
@@ -31,13 +32,16 @@ module Codec.CBOR.Term
   ( Term(..)    -- :: *
   , encodeTerm  -- :: Term -> Encoding
   , decodeTerm  -- :: Decoder Term
+  , TermF(..)   -- :: *
```
bgamari, line 35: Can you add a Haddock section here to set it apart from `Term`?

infinity0:
@bgamari Thanks for the encouragement! Though I've been playing around with `recursion-schemes` in the mean time and now think its approach is generally superior.

If there were no other concerns, I'd suggest simply using that here, to auto-generate `TermF` and the related other things via TH. This would help to avoid orphan instances, which would happen if it were auto-generated in another package, like I'm doing locally for now. However it has a [decent amount of extra dependencies](https://hackage.haskell.org/package/recursion-schemes) (the usual ekmett stuff) which I'm not sure you'd want for this library.

What the autogenerated code looks like is:

```haskell
    makeBaseFunctor ''Term
  ======>
    data TermF r
      = {- pretty much same as defined in this PR -}
    type instance Base Term = TermF
    instance Recursive Term where
      -- project :: Term -> TermF Term
      project = {- straightforward case-matching with no recursion -}
    instance Corecursive Term where
      -- embed :: TermF -> Term Term
      embed = {- straightforward case-matching with no recursion -}
```

The `toTermF` function I implemented so far, behaves exactly the same as the `cata` method on class `Recursive` with a default implementation of `cata f = c where c = f . fmap c . project` where `project` is as auto-generated above.

So one way to bring this PR closer to `recursion-scheme`s approach would be to implement `project` and `embed` instead of `toTermF`. Then someone could easily define `recursive-schemes` instances by reusing these functions, however they would still be orphan instances. What do you think, how were you using your local definitions?

infinity0:
Closing for now, need to re-purpose my fork's master branch for other stuff, might revisit again later.

PR #247, opened 2020-09-03
I am looking for cborg-0.2.4.0. Hackage pointed me here.

CodiePP:
newest version on Hackage is 0.2.4.0, but the cabal file is still at 0.2.3.0.


CodiePP:
it's in https://github.com/well-typed/cborg/releases/tag/cborg-0.2.4.0
so this is actually wrong.

PR #249, opened 2020-10-22
add instances for these, strict

infinity0:
`strict` is a recently-revived library containing strict versions of some base types, which also includes strict versions of `these` - , @phadej and I have been maintaining it.

This PR contains the relevant `Serialise` instances for those types. We thought it was better to keep these instances here instead of inside `strict` itself, because `strict` is meant to be a light library with minimal dependencies. Recently `aeson` and `lens` have added instances there too, depending on `strict`.

### DIFF
--- a/serialise/src/Codec/Serialise/Class.hs
+++ b/serialise/src/Codec/Serialise/Class.hs
@@ -70,10 +70,12 @@ import qualified Data.ByteString.Lazy                as BS.Lazy
 import qualified Data.Map                            as Map
 import qualified Data.Sequence                       as Sequence
 import qualified Data.Set                            as Set
+import qualified Data.Strict                         as Strict
 import qualified Data.IntSet                         as IntSet
 import qualified Data.IntMap                         as IntMap
 import qualified Data.HashSet                        as HashSet
 import qualified Data.HashMap.Strict                 as HashMap
+import qualified Data.These                          as These
 import qualified Data.Tree                           as Tree
 import qualified Data.Primitive.ByteArray            as Prim
 import qualified Data.Vector                         as Vector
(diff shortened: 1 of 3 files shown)


infinity0:
I added an instance for `Data.Void` as well in the meantime.

serialise/src/Codec/Serialise/Class.hs
```diff
@@ -204,6 +207,13 @@ instance Serialise a => Serialise (NonEmpty.NonEmpty a) where
 --------------------------------------------------------------------------------
 -- Primitive and integral instances
 
+#if MIN_VERSION_base(4,8,0)
+-- | @since 0.2.3.1
+instance Serialise Void where
```
phadej, line 212: Depend conditionally on `void` package.

serialise/src/Codec/Serialise/Class.hs
```diff
@@ -850,6 +860,44 @@ instance (Serialise a, Serialise b) => Serialise (Either a b) where
                           return (Right x)
                   _ -> fail "unknown tag"
 
+-- | @since 0.2.3.1
```
phadej, line 863: Adding instances requires at least minor bump, so at very least `@since 0.2.4.0`.

serialise/src/Codec/Serialise/Class.hs
infinity0: I don't know how to make sense of that, can you clarify? There are no unconditional dependencies in the cabal file at the moment, and this conditional definition is similar to the other ones in the same file - i.e. it doesn't exist if the version of base being used doesn't support it. By contrast the `void` package seems to be designed as an unconditional dependency that shims in support for older `base`, but I don't know if the maintainers want to do that.

phadej: https://oleg.fi/gists/posts/2019-06-03-compat-packages.html#s:2

infinity0: OK that makes sense now, so I would do something like if impl(ghc < 7.10) build-depends: void` in the cabal file and drop the `MIN_VERSION_base(4,8,0)` check for `Data.Void` in `Serialise.Class`.

Could one of the cborg maintainers confirm that you want this? I note the current approach of this package doesn't do this for e.g. `Data.Semigroup` - there is no attempt to shim in `semigroups` for older base.

phadej: I'd say the `semigroups` should also be shimmed, `serialise` depends on `semigroups` transitively anyways.

phadej:
ping @thoughtpolice

infinity0:
ping @bgamari

infinity0:
ping someone? :)

dcoutts:
Sadly this is the classic "where should the instances live" problem.

I'm not convinced they should live here. Adding deps on strict and these pulls in quite a few new dependencies.

phadej:
@dcoutts does it. Exactly, `these`, `strict` (and `assoc`). No more.

phadej:
Dependencies in master:

deps-master

With

```
package these
  flags: -assoc

package strict
  flags: -assoc
```

(flags added specifically for dependency purists)

deps

infinity0:
@dcoutts I am not too particularly bothered as long as the instances are *somewhere*, but was persuaded by @phadej 's arguments over in haskellari/strict#21 to have it here because we are trying (and have persuaded some other packages) to position `strict` (and together with `these` which is also small) as a fundamental and base-like library within the Haskell ecosystem.

Generally I have found it much easier and intuitive to fix space leaks by marking long-lived state strict in their datatype definitions, rather than trying to spend a long time chasing where I need to place a `seq` (which may be in third-party code, or related to lazy instances / lazy lens on a lazy-or-strict data structure that I am using strictly). I believe this is what we should be saying to Haskell newbies, and this should be The New Way for fixing space leaks in Haskell. `strict` is a necessary part of this general approach.

infinity0:
@dcoutts or any other maintainer - any new thoughts?

infinity0:
ping again - in case it wasn't clear, I'll mention the context again that in a similar situation, `aeson` agreed to depend on `strict` not the other way around. We are keeping the dependencies of `strict` very minimal.

bgamari:
I suppose adding `strict` and `these` dependencies won't be the end of the world. Sure.

PR #254, opened 2020-12-18
Adds support for ByteStrings in inside JSON

yaitskov:
I am implementing WebAuthn authentication.
Chrome browser returns part of data encoded in CBOR and uses 58/59 bytes (fixed byte string with length up to 255/65535 bytes) to encode key.

I think exposing such type as a hex string in JSON is best option.
Aeson.Array Number is much less efficient.

+ extra dependency on base16

### DIFF
--- a/cborg-json/cborg-json.cabal
+++ b/cborg-json/cborg-json.cabal
@@ -26,6 +26,7 @@ library
     base >=4.7 && < 4.15,
     aeson >=0.7 && <1.6,
     aeson-pretty >=0.8 && <0.9,
+    base16-bytestring >= 0.1 && <0.2,
     unordered-containers >=0.2 && <0.3,
     scientific >=0.3 && <0.4,
     text >=1.1 && <1.3,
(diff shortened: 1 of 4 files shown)


yaitskov:
@phadej

phadej:
I don't understand why I'm pinged here.

yaitskov:
I was looking for reviewers and found you among previous ones.

phadej:
That's the way I ended up "maintaining" plenty of stuff :)

yaitskov:
CI is green except on gch 7.8.4 because store package dropped 7.8.4 support
```
9883f44 - Remove configuration for building with ghc-7.8 (4 months ago) (Michael Sloan)
```

phadej:
could `store` maintainers correct the metadata then: https://matrix.hackage.haskell.org/#/package/store, there shouldn't be red cells.

phadej:
Please, let's not workaround other bugs (which are easily fixable). There is no hurry.

yaitskov:
PR CI is green, thanks to store team efforts.
https://github.com/mgsloan/store/issues/157

dcoutts (changes requested):
Yes, the CBOR/JSON support wehave should cover byte strings. That said, I think as a default we should follow the CBOR RFC as our spec. It says:

https://tools.ietf.org/html/rfc7049#section-4.1

>    o  A byte string (major type 2) that is not embedded in a tag that specifies a proposed encoding is encoded in base64url without padding and becomes a JSON string.

There are also tags for hex, but it sounds like your application just uses raw cbor bytestrings.

If that's ok for your application that's great. If not and you really want to use hex, then I think the thing we should think about is how to do cusomised non-default conversions that are not exactly the conversions that the RFC specifies.

PR #268, opened 2021-09-24
Fix imports when `OPTIMIZE_GMP` is off

georgefst:
It looks like this was broken by the CPP changes here: https://github.com/well-typed/cborg/pull/259/commits/821a031e81417e904928eb22964bf8cc29fd6998.

[Catching this in CI](https://github.com/well-typed/cborg/issues/207) would be nice.

I'm not confident this change is actually *exactly* what's needed in all cases, but it's enough to get things running on my RPi (`integer-simple`, `-optimize-gmp`), without breaking anything on my 64-bit Linux desktop (`+optimize-gmp`).


bgamari:
The same issue was fixed in #271.

PR #285, opened 2022-03-21
Replace haskell-ci with template from https://kodimensional.dev/github-actions

alexbiehl:
Seems like Haskell-CI doesn't support ghc-9.2.2. The template offered by https://kodimensional.dev/github-actions uses ghcup to set up GHC and Cabal.


FinleyMcIlwaine:
Haskell-CI added support for ghc-9.2.2

PR #298, opened 2022-08-22
Support GHC 9.4

parsonsmatt:
Fixes #291 

This is a *very* stupid PR, where I use `CPP` and then do `_f` in the 9.4 case to see what to put in. But the tests all pass, so 🎉 🤷🏻

PR #299, opened 2022-09-04
`cborg-json`: Support `aeson-2.1`

ocharles:


FinleyMcIlwaine:
Thanks! This is added in #298 🙂

PR #307, opened 2023-02-08
Magic.hs: Fix for machines without MEM_UNALIGNED_OPS  and base >= 4.17

kubek2k:


centromere:
@parsonsmatt Would you be willing to review and merge this if appropriate?

parsonsmatt:
I'm not a maintainer for this package

kubek2k:
Looks like it's rather @bgamari - could you have a look?

brprice:
FTR, my understanding is that this is only necessary for arches without `MEM_UNALIGNED_OPS`, i.e. anything other than `i386` and `x86_64` (and potentially for ghc<7.10) (in particular, `aarch64`/apple silicon). In particular, for x86_64, cborg-0.2.8.0 builds fine with ghc-9.4.3 -- I believe the cpp means that the definition on [Magic.hs:171](https://github.com/well-typed/cborg/blob/master/cborg/src/Codec/CBOR/Magic.hs#L171) is in effect. But it does fail to build on `aarch64-darwin`.

Perhaps to avoid confusion in the future the PR title and commit message could be updated to reflect this?

gelisam:
I independently came up with a [mostly-identical patch](https://github.com/well-typed/cborg/issues/311#issuecomment-1508989466), so I also think that this is the correct fix.

Bodigrim:
@FinleyMcIlwaine could this please be merged? Otherwise `cborg` does not work on M1 machines.

unorsk:
Having this problem on M2 too.

nicolashery:
We also had this problem building on Apple Silicon. Confirm that this PR fixed it. Looking forward to seeing it merged!

PR #308, opened 2023-03-01
Bump primitive upper bound

shlevy:
From https://hackage.haskell.org/package/primitive-0.8.0.0/changelog, the major version increase is due to a role annotation change to `PrimArray`, which isn't used in the repo.

Would be great to get a hackage revision or new release with this change.


bgamari:
I have pushed a revision with this change.

PR #314, opened 2023-04-26
Add SlicedByteArray to roundtrip tests and merge Fix

Saizan:


Saizan:
Tests generally fail or segfault without the merged fix.

PR #315, opened 2023-04-27
Support bytestring in json as Base64Url

Saizan:
Implementing changes requested on PR #254


cborg-json/tests/Main.hs
```diff
@@ -0,0 +1,27 @@
...
+
+tests :: TestTree
+tests =
+  testGroup "CBOR-JSON"
+    [ testGroup "unit tests"
+        [ testCase "decode variable ByteString as Base62Url String" $
+            Right ("", String "MDEy") @=? deserialiseFromBytes (decodeValue True)
+                                              (fromStrict . either error id $ HEX.decode "5803303132")
+        ]
+    ]
```
Saizan, line 27: Ideally we would test against Reference.termToJson.

Issue #318, opened 2023-05-12
Encoding allocates too many blocks in old generation


adamgundry:
I've been debugging an issue in a server application that stores a large (multi-gigabyte) state in memory and occasionally serialises it to disk using `serialise`. In principle this serialisation should be able to happen in constant space. However, when it does so, memory usage grows significantly and forces major GCs (which we want to avoid for performance reasons, as we're using the copying collector).

The following is an attempt to reproduce the issue (using only `cborg`). The script constructs some fully-evaluated data in memory, then serialises it to disk. During the serialisation stage (between the second and third vertical bars in the first graph) we see that the live bytes remains constant (i.e. unnecessary data is not being retained on the heap), but the total blocks size grows significantly. By looking at the `+RTS -Dg` output it appears that the growth is due to blocks going straight into generation 1, and hence not being freed until a major GC. I would expect instead that blocks allocated during serialisation rapidly become garbage, and hence stay in the nursery or generation 0 to be freed in a minor GC. Perhaps [eager promotion](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/rts/storage/gc/eager-promotion) might be kicking in somewhere?

This reproduction applies to GHC 9.6.1, `cborg` 0.2.8.0 and `bytestring` 0.11.4.0, although I've also seen it on earlier versions. It's not obvious to me whether this is an issue in one of the libraries or GHC itself, but I found a related issue in `serialise` (#317) so this seemed like a reasonable place to report it.

Curiously this seems to depend on the structure of the state type. The problematic behaviour occurs if the state is a pair of lists of unit, but if the state is a single list of unit (using the commented-out definitions of `mkState` and `encode`) then we get the expected behaviour, where the blocks size remains constant, as shown in the second graph. Exactly which state types work seems to vary somewhat with GHC/library versions and other details of the surrounding code.

Thanks to @bgamari and @mpickering for helping me get this far. Any further insights would be very welcome. :grin:

## Unexpected behaviour

![Problematic block size growth](https://github.com/well-typed/cborg/assets/1153317/e418caf1-ebdd-4b78-817c-c558f3f2856b)

## Expected behaviour

![Expected behaviour](https://github.com/well-typed/cborg/assets/1153317/c4c2bbc5-ba78-4fa9-b5c4-053b05a685d7)

## Reproducer

```hs
import Control.DeepSeq
import Debug.Trace
import System.Environment
import System.Mem

import qualified Codec.CBOR.Encoding as CBOR
[...]

PR #319, opened 2023-05-18
WIP: experimentally use dupIO

adamgundry:
See #318. This adds a new token type `TkDup` for use when encoding, that duplicates the contained closure. This means that encodings can explicitly duplicate thunks that may otherwise end up in the old generation.  It isn't obvious to me whether this is enough or whether calling `dupIO` every time would be better (and whether that might have an unreasonable performance cost).

Of course `dupIO` isn't yet released, and currently has soundness issues with static closures, so this isn't suitable for merge in any case...


adamgundry:
Closing in favour of #321.

PR #328, opened 2023-10-31
Make it build with ghc-9.8 and update dependencies

erikd:
Main compiler related issue was the absence of a `MonadFail Gen` instance.

Lot warnings from latest GHC.


binary-serialise-cbor/binary-serialise-cbor.cabal
```diff
@@ -51,7 +51,7 @@ library
   exposed-modules: Data.Binary.Serialise.CBOR.Read
 
   build-depends:
-    base       >= 4.11 && < 4.19,
+    base       >= 4.11 && < 4.21,
```
andreasabel, line 54: Why `< 4.21` and not `< 4.20`?  Is there evidence that it works with GHC-9.10?
If one follows a strict regime here, one should stick to the versions tested by CI.

andreasabel (commented):
LGTM, thanks!

`base < 4.21` is unusual, is this intended?  I would have expected `base < 4.20`.

binary-serialise-cbor/binary-serialise-cbor.cabal
erikd: `ghc-9.10` will use `base-4.21`.

`Base` skips odd numbers just like `ghc` skips odd numbers.

erikd: Fixed.

andreasabel (approved):

bgamari (commented):
Uploaded:
* `cborg-0.2.10.0`
* `binary-serialise-cbor-0.2.2.0-rev2`
* `cbor-tool-0.2.2.0-rev5`
* `cborg-json-0.2.6.0`
* `serialise-0.2.6.1`

Issue #343, opened 2025-03-04
`decodeValue` does not accept `TypeMapLen`


Saizan:
`decodeValue` from `cborg-json` does not accept indefinite-length maps for objects. The spec doesn't seem to rule it out as a valid encoding.

PR #345, opened 2025-04-09
cborg-json: support indefinite length maps when decoding objects

Saizan:
The CBOR<->JSON encoding spec does not rule out indefinite length maps, and some other implementations can produce them at times (e.g. rust's serde with flatten attributes). 

- [x] Test case.


Saizan:
This fixes issue #343 , not sure why the CI jobs get canceled though.

coot (approved):
LGTM

PR #349, opened 2025-07-05
Use dupIO to avoid unwanted sharing when encoding (another take)

phadej:
Rebased https://github.com/well-typed/cborg/pull/321 

@adamgundry asked "This raises the question of how to test it in CI..", `haskell-ci` has support for that.

I won't merge this as this adds a non-trivial dependency. Currently GHC-9.12 is excluded CI as `dupIO` doesn't allow it.


phadej:
It seems GHC-9.10 has to be excluded as well as GHC-9.10.2 major-bumped the `ghc-prim` version

phadej:
ping @adamgundry

PR #351, opened 2025-07-05
Non-x86_64 CI setup

phadej:


phadej:
Ok, so 32bit build is indeed broken:

```
src/Codec/CBOR/Decoding.hs:986:22: error: [GHC-83865]
    • Couldn't match expected type ‘Int#’ with actual type ‘Int64#’
    • In the first argument of ‘intToInt64#’, namely ‘off#’
      In the first argument of ‘I64#’, namely ‘(intToInt64# off#)’
      In the first argument of ‘k’, namely ‘(I64# (intToInt64# off#))’
```

I'll cleanup the CI setup itself, and then proceed with fixing the 32bit build.

PR #371, opened 2025-11-10
Support indefinite length bytestring

### LINKED ISSUE #263: Bignum decoding doesn't allow indefinite-length bytestrings
`readBigUInt` and `readBigNInt` assume (https://github.com/well-typed/cborg/blob/master/cborg/src/Codec/CBOR/Read.hs#L2660) that the payload is a definite-length bytestring, although as far as I can see the specification merely says it his a "byte string data item", which I think means both definite and indefinite length bytestrings should be allowed.

Example (bigint tag, and then a random indefinite-length bytestring I got the library to make for me):
```
> let y :: Term = deserialise $ BSL.fromStrict (BS.pack [0xc2, 0x5f, 0x43, 0x6c, 0x6f, 0x6c, 0xff])
> y
*** Exception: DeserialiseFailure 0 "expected integer"
```

qnikst:
We support indefinite length bytestring in the Read module.

In order to do it we add two extra variants for bigint deseralization, both puts us to the slow-path to start consumption of the chunks.

This serialization is considered non-canonical thus canonical read rejects is. While specification is silent about it, the cbor.me implementation always converts bigit into a fixed size byte blob.

Fixes #263.

### DIFF
--- a/cborg/cborg.cabal
+++ b/cborg/cborg.cabal
@@ -135,6 +135,7 @@ test-suite cborg-tests
     Tests.Regress
     Tests.Regress.Issue160
     Tests.Regress.Issue162
+    Tests.Regress.Issue263
     Tests.Regress.FlatTerm
     Tests.Reference
     Tests.Reference.Implementation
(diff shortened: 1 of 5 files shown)


coot:
There's a subtle bug in this implementation, which I fixed in #384.

PR #372, opened 2025-11-17
chore: upgrade QuickCheck dependencies bounds

blackheaven:


serialise/serialise.cabal
```diff
@@ -123,7 +123,7 @@ test-suite tests
     unordered-containers,
     cborg,
     serialise,
-    QuickCheck              >= 2.15    && < 2.17,
+    QuickCheck              >= 2.9     && < 2.18,
```
dcoutts, line 126: There's no need to relax the lower bound as far as I can see. The constraints for aeson force us to use QC 9.15 as a lower bound anyway.

cborg/cborg.cabal
```diff
@@ -155,7 +155,7 @@ test-suite cborg-tests
     base16-bytestring       >= 1.0     && < 1.1,
     deepseq,
     half,
-    QuickCheck              >= 2.9     && < 2.17,
+    QuickCheck              >= 2.9     && < 2.18,
```
dcoutts, line 158: You got there first but I ended up merging MR #378 first which also made this change.

dcoutts (approved):
Thanks @blackheaven! This is fine, but since #378 got merged, there's nothing left to do here.

PR #378, opened 2026-05-13
Upgrade to GHC 9.14

### LINKED ISSUE #373: Relax upper bounds for GHC 9.14.1
`serialise-0.2.6.1` builds with GHC 9.14.1 and its boot packages `base-4.22.0.0`, `containers-0.8` and `time-1.15`.

`cborg-0.2.10.0` builds with GHC 9.14.1 and its boot packages `base-4.22.0.0` and `containers-0.8`.

dancewithheart:
Fix #373

Tested locally:
```sh
cabal build all
cabal test all
cabal build all -w ghc-9.14.1
```

### DIFF
--- a/cborg/cborg.cabal
+++ b/cborg/cborg.cabal
@@ -21,7 +21,8 @@ tested-with:
   GHC == 9.6.7,
   GHC == 9.8.4,
   GHC == 9.10.2,
-  GHC == 9.12.2
+  GHC == 9.12.2,
+  GHC == 9.14.1
 
 extra-source-files:
   ChangeLog.md
--- a/cborg/cborg.cabal
+++ b/cborg/cborg.cabal
@@ -90,7 +90,7 @@ library
 
   build-depends:
     array                   >= 0.4     && < 0.6,
-    base                    >= 4.13    && < 4.22,
+    base                    >= 4.13    && < 4.23,
     bytestring              >= 0.10.4  && < 0.13,
     deepseq                 >= 1.4.4.0 && < 1.6,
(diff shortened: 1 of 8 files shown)


cborg/cborg.cabal
```diff
@@ -102,7 +103,7 @@ library
 
   if flag(optimize-gmp)
     cpp-options:            -DOPTIMIZE_GMP
-    if impl(ghc >= 9.0)
+    if impl(ghc >= 9.0 && < 9.14)
```
dcoutts, line 106: What is this bit about? Why can't we use ghc-bignum with ghc-9.14?

dcoutts (commented):
LGTM! Just one query. I'll try seeing if we can avoid the bignum change.

cborg/cborg.cabal
dcoutts: Ok, I've built it and I see the problem. But the fix should be different. I've made a fix and I'll merge with that extra patch.

dcoutts:
Thanks @dancewithheart !

cborg/cborg.cabal
dancewithheart: Thank you for taking care of this. I was just following the errors.

BTW I really enjoyed your talks about performance and Cardano 🙂

andreasabel:
It seems that bumping `time` has been forgotten.
- #381
 
Base package version mismatch is caught by Haskell CI, which is consequently broken atm.
@dancewithheart @dcoutts The failure of CI should maybe have been taken seriously: https://github.com/well-typed/cborg/actions/runs/25780222867/job/75801264002

